PR #48, opened 2020-12-13
stack was implemented

Mandybur800:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,63 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> head;
+    private Node<T> tail;
```
IvanLuchkin, line 7: Since the `Node` only has one reference to the other `Node`, we only need either `head` or `tail`, having both is redundant

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,63 @@
...
+    public void push(T value) {
+        if (size == 0) {
+            head = new Node<>(value, null);
+            tail = head;
+            size++;
+            return;
+        }
+        tail.next = new Node<>(value, null);
+        tail = tail.next;
+        size++;
```
IvanLuchkin, line 23: This can be simplified if we imagine our stack as a singly list (see task description and previous comment)

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,63 @@
...
+        checkStack();
+        Node<T> localNode = head;
+        size--;
+        for (int i = 0; i < size - 1; i++) {
+            localNode = localNode.next;
+        }
+        T result = tail.value;
+        tail = localNode;
+        tail.next = null;
+        return result;
```
IvanLuchkin, line 41: We don't need to iterate since any stack implementation is LIFO data structure, we are always taking the top-most element

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,63 @@
...
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+
+        public Node(T value, Node<T> next) {
```
IvanLuchkin, line 58: I don't think that making this constructor `public` is a good idea. I would recomment changing the access modifier to `default`.

IvanLuchkin (changes requested):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
+    private int size;
 
+    public MateStack() {
+        size = 0;
+        tail = new Node<>(null,null);
```
IvanLuchkin, line 11: Your stack will always have some extra node which does not contain any value. You can dismiss initializing `tail` because if it is not initialized, it will be null by defaulf (line 6).

IvanLuchkin (changes requested):

IvanLuchkin (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
+    private int size;
 
+    public MateStack() {
+        size = 0;
```
Sofasmile, line 10: it can be removed

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
```
Sofasmile, line 6: better rename `tail` to `top`  or `head`

Sofasmile (approved):

PR #49, opened 2020-12-15
Stack by oleh kravets

Ketch92:


src/main/java/core/stack/MateStack.java
```diff
@@ -18,10 +20,16 @@ public void push(T value) {
...
+        if (isEmpty()) {
+            throw new EmptyStackException();
+        }
         return stack[size - 1];
     }
 
     public T pop() {
+        if (isEmpty()) {
+            throw new EmptyStackException();
+        }
```
gigbat, line 32: You can create a separate method for this part. Because you call the same in another method.

src/main/java/core/stack/MateStack.java
```diff
@@ -18,10 +20,16 @@ public void push(T value) {
...
         T value = stack[size - 1];
         stack[size-- - 1] = null;
```
gigbat: Strange decided, i hope mentor will like it :)

gigbat (changes requested):

src/main/java/core/stack/MateStack.java
Ketch92: I know, but I like it)

gigbat (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
...
-        return null;
+        validatePeekAndPop();
+        return stack[size - 1];
     }
 
     public T pop() {
-        return null;
+        validatePeekAndPop();
+        T value = stack[size - 1];
+        stack[size-- - 1] = null;
```
Resci, line 30:
```suggestion
        stack[--size] = null;
```

Resci (changes requested):

Resci (approved):

KseniiaMakarova (approved):
great)

PR #50, opened 2020-12-15
Task completed

NikolayFatkullin:
- Implemented methods push(), peek(), pop(), size();
- Created inner class Node<V>;
- Created method isEmpty();

### DIFF
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -1,19 +1,52 @@
 package core.stack;
 
-public class MateStack<T> {
-    public void push(T value) {
+import java.util.EmptyStackException;
 
+public class MateStack<T> {
+    private int size;
+    private Node<T> tail;
+
+    public void push(T value) {
+        Node<T> newNode = new Node<>(value, tail);
+        newNode.previous = tail;
+        tail = newNode;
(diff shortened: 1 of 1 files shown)


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private int size;
+    private Node<T> tail;
 
+    public void push(T value) {
+        Node<T> newNode = new Node<>(value, tail);
+        newNode.previous = tail;
```
dmytro-hryhoruk, line 11: you already doing this in the constructor

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private static class Node<V> {
+        V value;
+        Node<V> previous;
```
dmytro-hryhoruk, line 36: access modifiers

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

PR #53, opened 2020-12-16
Implemented all methods in class 'MateStack'

YuraMitryakhin:


Vedroid (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
     }
 
     public T pop() {
-        return null;
+        checkSize();
+        Node<T> removeNode = tail;
+        tail = removeNode.prevNode;
+        size--;
+        return removeNode.value;
+
```
dmytro-hryhoruk, line 26:
```suggestion
```

dmytro-hryhoruk (approved):

PR #55, opened 2020-12-19
Stack implemented

Artificial-Friend:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+            this.current = new Node<>(current, value);
+        }
+        ++size;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
Vedroid, line 22: Move to a separate method

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+        return current.value;
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
Vedroid, line 29: Move to a separate method

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+    private int size;
+    private Node<T> current;
 
+    public void push(T value) {
+        if (size == 0) {
+            current = new Node<>(null, value);
+        } else {
+            Node<T> current = this.current;
+            this.current = new Node<>(current, value);
+        }
```
Vedroid, line 15:
```suggestion
        current = new Node<>(current, value);
```

Vedroid (changes requested):

src/main/java/core/stack/MateStack.java
dmytro-hryhoruk: agree and let's check the size before this

src/main/java/core/stack/MateStack.java
dmytro-hryhoruk: agree

dmytro-hryhoruk (changes requested):

src/main/java/core/stack/MateStack.java
Vedroid: Why do we need to check the size here?

dmytro-hryhoruk (approved):

PR #57, opened 2020-12-20
created stack via array

YuraBondarenko:


antoniukivan (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+    private static final int DEFAULT_STACK_SIZE = 10;
+    private T[] stack;
+    private int size;
+
+    public MateStack() {
+        stack = (T[]) new Object[DEFAULT_STACK_SIZE];
+    }
 
+    public void push(T value) {
+        stack[size] = value;
```
KseniiaMakarova, line 15: what if `size` is 10 or more?

KseniiaMakarova (changes requested):

src/main/java/core/stack/MateStack.java
```diff
@@ -45,4 +47,12 @@ private T getElement() {
         isEmptyStack();
         return stack[getIndexOfLastElement()];
     }
+
+    private boolean resize() {
```
KseniiaMakarova, line 51: do you use this return type?

src/main/java/core/stack/MateStack.java
```diff
@@ -45,4 +47,12 @@ private T getElement() {
...
+        if (size == stack.length) {
+            stack = Arrays.copyOf(stack, (int) (stack.length * 1.5));
```
KseniiaMakarova, line 53: let's make 1.5 a constant as well

KseniiaMakarova (changes requested):

src/main/java/core/stack/MateStack.java
YuraBondarenko: Bogdan said that better not to use void methods because it is harder to test

KseniiaMakarova: will you test private methods?

KseniiaMakarova (approved):

PR #60, opened 2020-12-21
add solution

antoniukivan:


YuraBondarenko (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private static class Node<T> {
+        private final T value;
+        private final Node<T> previous;
```
KseniiaMakarova, line 39: why final?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> previous;
```
KseniiaMakarova, line 6:
```suggestion
    private Node<T> top;
```
what do you think?

KseniiaMakarova (changes requested):
nice work!

src/main/java/core/stack/MateStack.java
antoniukivan: i didn't want, it's all compiler))

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> top;
```
KseniiaMakarova, line 39:
```suggestion
        private Node<T> previous;
```
`top` is the top node in the stack, but each node has the link to the previous one, right?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
-        return null;
+        checkSize();
+        return top.value;
     }
 
     public T pop() {
-        return null;
+        checkSize();
+        T value = top.value;
+        top = top.top;
```
KseniiaMakarova, line 22:
```suggestion
        top = top.previous;
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #61, opened 2020-12-21
MateStack class implemented

semyonich:


dmytro-hryhoruk (changes requested):
image

dmytro-hryhoruk (changes requested):
image

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+        return elementData[size - 1];
     }
 
     public T pop() {
-        return null;
+        if (size() == 0) {
+            throw new EmptyStackException();
+        }
+        size--;
+        return elementData[size];
```
dmytro-hryhoruk, line 35: this object will still remain in the array. Try to use System.arraycopy for this

dmytro-hryhoruk (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int INITIAL_LENGTH = 10;
+
```
dmytro-hryhoruk, line 7:
```suggestion
```

dmytro-hryhoruk (changes requested):

src/main/java/core/stack/MateStack.java
semyonich: done

src/main/java/core/stack/MateStack.java
semyonich: done

dmytro-hryhoruk (approved):

PR #63, opened 2020-12-21
added implementation for four main methods, inner class Node and one …

Fedich09:
…helper methods

### DIFF
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -1,19 +1,81 @@
 package core.stack;
 
-public class MateStack<T> {
-    public void push(T value) {
+import java.util.EmptyStackException;
 
+public class MateStack<T> {
+    private Node<T> head;
+    private int size;
+
+    public void push(T value) {
+        if (head == null) {
+            head = new Node<>(null, value);
+        } else {
+            Node<T> current = head;
+            while (true) {
+                if (current.next == null) {
+                    current.next = new Node<>(null, value);
+                    break;
+                }
+                current = current.next;
+            }
+        }
+        size++;
     }
 
     public T peek() {
-        return null;
+        sizeChecker(size);
+        Node<T> current = head;
+        while (true) {
+            if (current.next == null) {
+                return current.item;
+            }
+            current = current.next;
+        }
     }
 
     public T pop() {
-        return null;
+        sizeChecker(size);
+        T retValue;
+        Node<T> last = head;
+        while (true) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,81 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> head;
```
NikolayFatkullin, line 6: In Stack we use LIFO (Last In First Out), so for optimization, the better way Is use the tail of your Node lists, but it is just advice to you =) And if you remake your code with using tail, your code will  smaller

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,81 @@
...
+
+    private void sizeChecker(int size) {
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    private static class Node<T> {
+        Node<T> next;
+        T item;
```
NikolayFatkullin, line 74: access modifiers 
```suggestion
        private Node<T> next;
        private T item;
```

NikolayFatkullin (changes requested):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
+    private int size;
 
+    public void push(T value) {
+        if (tail == null) {
+            tail = new Node<>(null, value);
+        } else {
```
NikolayFatkullin, line 12: this if statement is not necessary =) You can delete it =)

NikolayFatkullin (changes requested):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
+        size--;
+        return retValue;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private void sizeChecker(int size) {
```
dmytro-hryhoruk, line 31:
```suggestion
    private void checkSize(int size) {
```
sizeChecker it's more likely the name of the class than a method

dmytro-hryhoruk: method names should always start with verb

dmytro-hryhoruk (approved):

PR #64, opened 2020-12-21
added solution stack

IgorBochkor:
added solution stack

### DIFF
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -1,19 +1,51 @@
 package core.stack;
 
-public class MateStack<T> {
-    public void push(T value) {
+import java.util.EmptyStackException;
 
+public class MateStack<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] array;
+    private int size;
+
+    public MateStack() {
+        array = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
+    private void grow() {
+        Object[] newArray = new Object[array.length + (array.length >> 1)];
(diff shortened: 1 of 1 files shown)


antoniukivan (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] array;
+    private int size;
 
+    public MateStack() {
+        array = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
+    private void grow() {
```
KseniiaMakarova, line 15: where should private methods be located?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+        checkException();
+        return array[size - 1];
     }
 
     public T pop() {
-        return null;
+        checkException();
+        T element = array[size - 1];
+        System.arraycopy(array, size + 1, array, size - 1, 0);
+        size--;
```
KseniiaMakarova, line 38: how many elements do you copy on line 37?

let's just make the last element null instead

KseniiaMakarova (changes requested):

src/main/java/core/stack/MateStack.java
IgorBochkor: after public
fixed

src/main/java/core/stack/MateStack.java
IgorBochkor: done

KseniiaMakarova (approved):

PR #65, opened 2020-12-22
first commit

RuslanIbIsh:


oleksandrdziubiuk (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
-        return null;
+        checkStack();
+        return (T) stack[size - 1];
     }
 
     public T pop() {
-        return null;
+        checkStack();
+        T result = (T) stack[size - 1];
+        size--;
```
KseniiaMakarova, line 32: it will also be good to make the element on `stack[size - 1]` null before we decrement `size`

KseniiaMakarova (approved):

PR #67, opened 2020-12-22
Added implementation

sanek98cv:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,45 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
+    private int size;
 
+    public MateStack() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
+    public void push(T value) {
+        elementData[size++] = value;
```
RuslanIbIsh, line 15: what if indexOutOfBound?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,45 @@
...
     }
 
     public T peek() {
-        return null;
+        if (size() == 0) {
+            throw new EmptyStackException();
```
RuslanIbIsh, line 20: you can move this check to the separate method and use it throw all code

RuslanIbIsh (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private void resize() {
+        elementData = Arrays.copyOf(elementData,
+                DEFAULT_CAPACITY * RESIZE_ELEMENT_DATA);
```
KseniiaMakarova, line 40: what if we will need to resize several times? can DEFAULT_CAPACITY change?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
+            throw new ArrayIndexOutOfBoundsException("Index incorrect");
+        }
+        if (size - index - 1 > 0) {
+            System.arraycopy(elementData, index + 1,
+                    elementData, index, size - index - 1);
+        }
+        elementData[size--] = null;
+    }
+
+    private void getCheckSize() {
```
KseniiaMakarova, line 54:
```suggestion
    private void checkSize() {
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
+        if (size() == 0) {
```
KseniiaMakarova, line 55: can we use field instead of `size()`?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
     public T peek() {
-        return null;
+        getCheckSize();
+        return (T) elementData[size - 1];
     }
 
     public T pop() {
-        return null;
+        T value = peek();
+        removeElement(size() - 1);
```
KseniiaMakarova, line 30: same

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
+
+    private void resize() {
+        elementData = Arrays.copyOf(elementData,
+                DEFAULT_CAPACITY * RESIZE_ELEMENT_DATA);
+    }
+
+    private void removeElement(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("Index incorrect");
+        }
```
KseniiaMakarova, line 46: are you sure that you need this check? you will pass `size - 1` as index to this method

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
+        if (size - index - 1 > 0) {
+            System.arraycopy(elementData, index + 1,
+                    elementData, index, size - index - 1);
+        }
```
KseniiaMakarova, line 50: same here - is it necessary?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
+        elementData[size--] = null;
```
KseniiaMakarova, line 51: let's just make the last element null
what will be the index of the last element?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #68, opened 2020-12-22
Stack

Baldinini:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+        public Node(Node<T> prev, T value) {
+            this.prev = prev;
+            this.value = value;
+        }
+    }
+
+    private boolean stackIsEmpty() {
+        if (size != 0) {
+            return true;
+        }
```
antoniukivan, line 46: your method returns true if stack isn't empty , let's rename it (stackIsNotEmpty, checkSize, checkIfNotEmpty)

antoniukivan (changes requested):

antoniukivan (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+        private T value;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T value) {
+            this.prev = prev;
+            this.value = value;
+        }
+    }
+
+    private boolean stackIsNotEmpty() {
```
KseniiaMakarova, line 43:
```suggestion
    private void checkStackSize() {
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+    private int size;
 
+    public void push(T value) {
+        tail = new Node<>(tail, value);
+        size++;
     }
 
     public T peek() {
-        return null;
+        return stackIsNotEmpty() ? tail.value : null;
```
KseniiaMakarova, line 15:
```suggestion
        checkStackSize();
        return tail.value;
```

the problem is that you will never return null - this is a bit confusing

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #69, opened 2020-12-22
first commit

entertainment-veks:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail = null;
```
NikolayFatkullin, line 6: You don't need to declare tail here with null
```suggestion
    private Node<T> tail;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+    private int size;
 
+    public void push(T value) {
+        Node<T> current = new Node<>(value, tail);
+        tail = current; //we can inline but it will be less readable
```
NikolayFatkullin, line 11: delete comment
```suggestion
        tail = current;
```

NikolayFatkullin (changes requested):

NikolayFatkullin (approved):

dmytro-hryhoruk (approved):

PR #71, opened 2020-12-22
all methods done

vlykhvar:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
+
+    private void emptyStack() {
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    private static class Node<T> {
+        T value;
+        Node next;
```
Ketch92, line 39: missing access modifiers

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private int size = 0;
```
Ketch92, line 6: no need to initialize it

Ketch92 (changes requested):

src/main/java/core/stack/MateStack.java
vlykhvar: thks. has changed

src/main/java/core/stack/MateStack.java
vlykhvar: changed

Ketch92 (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node next;
+
+        public Node(T value, Node<T> next) {
```
KseniiaMakarova, line 41: is the constructor available from outside if the Node class is private?
```suggestion
        private Node(T value, Node<T> next) {
```

KseniiaMakarova (approved):

PR #73, opened 2020-12-22
added Stack methods implementation

oroboro-cpu:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final int CAPACITY_MULTIPLY = 2;
```
YuraBondarenko, line 7: I am not sure if it necessarily, but increase might be 1.5 as in ArrayList

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+    private T[] array;
+    private int capacity;
+
+    public MateStack() {
+        array = (T[]) new Object[INITIAL_CAPACITY];
+        capacity = 0;
```
YuraBondarenko, line 13: you can remove this line

YuraBondarenko (changes requested):

src/main/java/core/stack/MateStack.java
oroboro-cpu: Vector documentation states that the capacity is doubled each time it needs to grow

YuraBondarenko: ok, thanks, I'll know)

YuraBondarenko (approved):

KseniiaMakarova (commented):
please write your solution in a new branch - check out `git reset --soft` command for example, it will help to remove the commit from master branch

KseniiaMakarova (changes requested):
everything else is good - nice job!

KseniiaMakarova (changes requested):
new branch was not created

PR #74, opened 2020-12-22
Implemented methods :

Likh-Alex:
`push`, `peek`, `pop`, `size`

### DIFF
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -8,7 +8,7 @@ public class MateStack<T> {
     private int size = 0;
 
     public MateStack() {
-        this.stack = (T[]) new Object[DEFAULT_CAPACITY];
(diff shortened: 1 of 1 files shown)


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+        resize();
+        stack[size] = value;
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
vlykhvar, line 23: how about to make new method checkEmptyStack() and use them here and in pop() as well?

Likh-Alex: Done, made a separate private method `checkEmptyStack()`

vlykhvar (approved):

KseniiaMakarova (approved):

PR #75, opened 2020-12-22
added fields: CAPACITY, COEFFICIENT, NUMBER_ONE, array, size

ShemsheyMaxim:
added logic to methods: push(), peek(), pop(), size()
added method resize().

### DIFF
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -1,19 +1,51 @@
 package core.stack;
 
-public class MateStack<T> {
-    public void push(T value) {
+import java.util.EmptyStackException;
 
+public class MateStack<T> {
+    private static final int CAPACITY = 10;
+    private static final double COEFFICIENT = 0.5;
+    private static final int NUMBER_ONE = 1;
+    private T[] array;
+    private int size;
+
+    public MateStack() {
+        array = (T[]) new Object[CAPACITY];
+        size = 0;
+    }
+
+    public void push(T value) {
+        if (size >= array.length) {
+            array = resize(array);
+        }
+        array[size] = value;
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        return array[size - NUMBER_ONE];
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        T currentValue = array[size - NUMBER_ONE];
+        size--;
+        array[size] = null;
+        return currentValue;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private T[] resize(T[] array) {
+        T[] newArray = (T[]) new Object[(int) (array.length * COEFFICIENT + array.length)];
+        System.arraycopy(array, 0, newArray, 0, array.length);
+        return newArray;
     }
 }


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+    private int size;
 
+    public MateStack() {
+        array = (T[]) new Object[CAPACITY];
+        size = 0;
+    }
+
+    public void push(T value) {
+        if (size >= array.length) {
+            array = resize(array);
```
Fedich09, line 19:
```suggestion
            resize();
```
we can make this method void, and just call him

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+    private T[] array;
+    private int size;
 
+    public MateStack() {
+        array = (T[]) new Object[CAPACITY];
+        size = 0;
+    }
+
+    public void push(T value) {
+        if (size >= array.length) {
```
Fedich09, line 18:
```suggestion
        if (size == array.length) {
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+        }
+        array[size] = value;
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
Fedich09, line 28:
```suggestion
        if (size == 0) {
            throw new EmptyStackException();
        }
```
add this in separate method please

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+        return array[size - NUMBER_ONE];
```
Fedich09, line 29: maybe just ONE ?)) NUBER_ONE look like little tricky 
```suggestion
        return array[size - NUMBER_ONE];
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
Fedich09, line 35:
```suggestion
        if (size == 0) {
            throw new EmptyStackException();
        }
```
add this in separate method please

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+        T currentValue = array[size - NUMBER_ONE];
+        size--;
+        array[size] = null;
```
Fedich09, line 38:
```suggestion
        T currentValue = array[size - NUMBER_ONE];
        size--;
        array[size - 1] = null;
```
we return size -  1 and after that we said that size will equal null, maybe size - 1 will equal null, I'm not sure about that

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private T[] resize(T[] array) {
+        T[] newArray = (T[]) new Object[(int) (array.length * COEFFICIENT + array.length)];
+        System.arraycopy(array, 0, newArray, 0, array.length);
+        return newArray;
```
Fedich09, line 49:
```suggestion
    private void resize() {
        T[] oldArray = array;
        array = (T[]) new Object[(int) (array.length * COEFFICIENT + array.length)];
        System.arraycopy(array, 0, oldArray, 0, oldArray.length);
```

Fedich09 (changes requested):

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: why void method better than return statement?

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: done

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: done

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: done

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: done

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: In this logic i decrease size and than use this size (which already - 1) and this work correctly. But I aggre with you and I changed line 37 and 38 places and use for understanding array[size - ONE] = null

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: why void method better than return statement?

src/main/java/core/stack/MateStack.java
Fedich09: less code )

src/main/java/core/stack/MateStack.java
Fedich09: less code)

src/main/java/core/stack/MateStack.java
Fedich09: understood )

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: done

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: done

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int CAPACITY = 10;
+    private static final double COEFFICIENT = 0.5;
+    private static final int ONE = 1;
```
dmytro-hryhoruk, line 8: okay, we may not make constants out of 1 and 0 numbers, they are not considered magic

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private void resize() {
+        T[] oldArray = array;
+        array = (T[]) new Object[(int) (array.length * COEFFICIENT + array.length)];
```
dmytro-hryhoruk, line 44: can we just multiple our length by 1.5?

dmytro-hryhoruk (changes requested):

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: done

src/main/java/core/stack/MateStack.java
ShemsheyMaxim: done

dmytro-hryhoruk (approved):

PR #78, opened 2020-12-22
Hw

engineer027:


sanek98cv (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> head;
```
KseniiaMakarova, line 6: do you need the head pointer?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
...
+    private Node<T> tail;
+    private int size = 0;
+
+    private static class Node<T> {
```
KseniiaMakarova, line 10: let's move this class to the bottom

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
...
+        private T value;
+        private Node<T> prev;
+
+        public Node(T value, Node<T> prev) {
```
KseniiaMakarova, line 14:
```suggestion
        private Node(T value, Node<T> prev) {
```
will this constructor be available from outside if the Node class is private?

KseniiaMakarova (changes requested):

src/main/java/core/stack/MateStack.java
engineer027: no

src/main/java/core/stack/MateStack.java
engineer027: +

src/main/java/core/stack/MateStack.java
engineer027: ok

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
+    private int size = 0;
 
+    public void push(T value) {
+        Node<T> newNode = new Node<>(value, null);
+        if (size != 0) {
+            newNode.prev = tail;
+        }
```
KseniiaMakarova, line 13: do you need this check? what will tail be if size is 0?

KseniiaMakarova (changes requested):

src/main/java/core/stack/MateStack.java
engineer027: ok

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
+    private int size = 0;
 
+    public void push(T value) {
+        Node<T> newNode = new Node<>(value, null);
+        newNode.prev = tail;
```
KseniiaMakarova, line 11:
```suggestion
        Node<T> newNode = new Node<>(value, tail);
```

KseniiaMakarova (approved):

PR #79, opened 2020-12-22
1) created custom Stack

AntonPanchishko:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> head;
```
Max-Pochepets, line 6: better call it head

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+    private int size;
 
+    public MateStack() {
+        this.size = 0;
```
Max-Pochepets, line 10: 0 is default value of int

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+    }
+
+    public void push(T value) {
+        Node<T> newNode = new Node<>(value, head);
+        head = newNode;
```
Max-Pochepets, line 15: it can be done in one line

Max-Pochepets (changes requested):

Max-Pochepets (approved):
👍

dmytro-hryhoruk (approved):

PR #80, opened 2020-12-22
made solution

Sashk52:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+    }
 
+    public void push(T value) {
+        Node newNode = new Node(null, value);
+        if (tail == null) {
+            tail = newNode;
+        } else {
+            newNode.previousNode = tail;
+            tail = newNode;
+        }
```
NikolayFatkullin, line 26: Do you really need this 'if' statement? And also, you can declare 'newNode.previousNode' in the constructor.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> previousNode;
+
+        public Node(Node previousNode, T value) {
```
NikolayFatkullin, line 13: Access modifier. You don't need public access for this class

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+        size--;
+        return (T) temp.value;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    public void isEmptyStack() {
```
NikolayFatkullin, line 47: Access modifier. Do you need public access to this method?

NikolayFatkullin (changes requested):

src/main/java/core/stack/MateStack.java
Sashk52: actually not, private good option. done.

src/main/java/core/stack/MateStack.java
Sashk52: agree, done.

src/main/java/core/stack/MateStack.java
Sashk52: You are rigth,anyway i make my Node tail and it not depends on exist tail or not. done

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
+        return tail.value;
     }
 
     public T pop() {
-        return null;
+        isEmptyStack();
+        Node temp = tail;
+        tail = tail.previousNode;
+        size--;
+        return (T) temp.value;
```
dmytro-hryhoruk, line 34:
```suggestion
        Node<T> temp = tail;
        tail = tail.previousNode;
        size--;
        return temp.value;
```

dmytro-hryhoruk (approved):

src/main/java/core/stack/MateStack.java
Sashk52: done

PR #81, opened 2020-12-22
created all methods in class MateStack

Alekearl:


engineer027 (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int INITIAL_SIZE_CAPACITY = 10;
+    private static final int ADDING_SIZE = 5;
+    private T[] array;
+    private int size;
 
+    public MateStack() {
+        this.array = (T[]) new Object[INITIAL_SIZE_CAPACITY];
```
KseniiaMakarova, line 12:
```suggestion
        array = (T[]) new Object[INITIAL_SIZE_CAPACITY];
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+        size--;
+        return value;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private T[] grow(T[] arrayValues) {
```
KseniiaMakarova, line 42:
```suggestion
    private void grow(T[] arrayValues) {
```
is an array changed in the whole application when it is modified inside some method?

KseniiaMakarova (changes requested):

src/main/java/core/stack/MateStack.java
Alekearl: I think if I change the state of the array in method, then the original array will be modified as well.

KseniiaMakarova (approved):

PR #82, opened 2020-12-22
stack implemented

less08:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,64 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
+    private int size;
+    private Node<T> head;
```
Max-Pochepets, line 7: better call it tail, IMHO

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,64 @@
...
+
+    private static class Node<T> {
```
Max-Pochepets, line 9: https://mate-academy.github.io/style-guides/java/java.html#s3.4.2-ordering-class-contents
better move it after public methods

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,64 @@
...
+        Node<T> current = head;
+        if (head == null) {
+            head = temp;
+        } else {
 
+            while (current.next != null) {
+                current = current.next;
+            }
+            current.next = temp;
+        }
```
Max-Pochepets, line 30: if u change head to tail, all of this can be done in one line

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,64 @@
...
 
     public T peek() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        Node<T> current = head;
+        while (current.next != null) {
+            current = current.next;
+        }
```
Max-Pochepets, line 41: same

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,64 @@
...
+        return current.value;
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
Max-Pochepets, line 48: you are using the same logic twice, maybe we should move it to separate method?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,64 @@
...
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
+    private int size;
+    private Node<T> head;
+
+    private static class Node<T> {
+        T value;
+        Node<T> next;
```
Max-Pochepets, line 11: maybe not next? if we are moving from the end in stack?

Max-Pochepets (changes requested):

src/main/java/core/stack/MateStack.java
dmytro-hryhoruk: up to you

src/main/java/core/stack/MateStack.java
dmytro-hryhoruk: agree

src/main/java/core/stack/MateStack.java
dmytro-hryhoruk: Okay, considering the code below, I would suggest you fix all @Max-Pochepets  comments. Push method Should not iterate through the whole stack to find the last one. You should keep your current node at the last position of your stack.

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

PR #84, opened 2020-12-23
initial commit

AleksandrPok:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+    private T returnValue() {
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        return array[size - 1];
+    }
+
+    private void resize() {
+        T[] bufferArray = array;
+        array = (T[]) new Object[array.length + (array.length >> 1)];
```
Max-Pochepets, line 46: what >> do?

Max-Pochepets (approved):

src/main/java/core/stack/MateStack.java
dmytro-hryhoruk: right shift divides the number by 2. So basically he is increasing array in 1.5 times.

dmytro-hryhoruk (approved):

src/main/java/core/stack/MateStack.java
Max-Pochepets: 👍

PR #86, opened 2020-12-23
Stack created

AndriiBardin:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> previous;
+
+        public Node(Node<T> previous, T value) {
```
KseniiaMakarova, line 41:
```suggestion
        private Node(Node<T> previous, T value) {
```
will the constructor be available from outside if the Node class is private?

KseniiaMakarova (approved):

PR #88, opened 2021-04-11
MateStack implementation added

ritakisluhina:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+        }
+    }
 
+    public void push(T value) {
+        if (size == 0) {
+            element = new Node<>(value, null);
+        } else {
+            element = new Node<>(value, element);
+        }
+        size++;
```
freedmanMain, line 25:
```suggestion
        Node<T> newNode = new Node<>(value, element);
        element = newNode;
        size++;
```
Can we not use (if and else) here ? We can do like this.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> element;
```
freedmanMain, line 6:
```suggestion
    private Node<T> top;
```
Let's rename this field to "top" or "upper".

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+        isEmpty();
+        Node<T> removed = element;
+        final T result = removed.item;
+        if (removed.next == null) {
+            element = null;
+        } else {
+            element = removed.next;
+        }
+        size--;
+        return result;
```
freedmanMain, line 43: Same.

ritakisluhina: done

src/main/java/core/stack/MateStack.java
ritakisluhina: done

src/main/java/core/stack/MateStack.java
ritakisluhina: I want to have as it is (:

freedmanMain (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> element;
```
RomanDubovskyi, line 6: element is to abstract, this field is not just random element it's last/top or smt like this, it should clear that this element is not from the middle or random place

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
+        size--;
+        return result;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    public void isEmpty() {
```
RomanDubovskyi, line 41: method with such naming usually returns boolean
```suggestion
    private void validateSize() {
```

RomanDubovskyi (changes requested):
good job overall

Max-Pochepets (approved):

PR #89, opened 2021-04-12
added solution

freedmanMain:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
...
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+
+        public Node(T value, Node<T> next) {
+            this.value = value;
+            this.next = next;
+        }
     }
 }
```
ritakisluhina, line 47: can we put this class upper so we can read code easier?
Good job!

ritakisluhina (changes requested):

src/main/java/core/stack/MateStack.java
ritakisluhina: Nice!

ritakisluhina (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
...
     public T peek() {
-        return null;
+        isEmptyStack();
+        return top.value;
     }
 
     public T pop() {
-        return null;
+        isEmptyStack();
+        T oldValue = top.value;
```
mankokolya, line 32: no need to call isEmtyStack() here and top.value;
you can just call oldValue = peek()
and as result you don't need method isEmptyStack(). you can check in peek

src/main/java/core/stack/MateStackMain.java
```diff
@@ -0,0 +1,13 @@
+package core.stack;
+
+public class MateStackMain {
```
mankokolya, line 3: You can remove this class from git

mankokolya (changes requested):

src/main/java/core/stack/MateStack.java
freedmanMain: Yes, that's true thanks.

mankokolya (approved):

RomanDubovskyi (approved):

RomanDubovskyi (approved):

RomanDubovskyi (approved):

PR #93, opened 2021-04-14
Stack task done

mankokolya:


freedmanMain (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,42 @@
...
+
+    public MateStack() {
+        container = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     public void push(T value) {
+        if (size == container.length) {
+            resize();
+        }
+        container[++size] = value;
```
boroda4436, line 19: On which position in the array you will add the first element?

mankokolya: Oops. At first position. Corrected

boroda4436 (approved):

PR #94, opened 2021-04-14
added solution

Dalai-S:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+
 public class MateStack<T> {
-    public void push(T value) {
+    private int size;
+    private Node<T> head;
+
+    static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> previous;
```
Nikber5, line 12: There is a condition in readme file: "Your implementation should be based on the array or singly linked list (don't use existing collections)". And your Linked list is double linked.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private int size;
+    private Node<T> head;
+
+    static class Node<T> {
```
Nikber5, line 9: I think you can make it private

Nikber5 (changes requested):

src/main/java/core/stack/MateStack.java
Dalai-S: Thank you, did not notice this information

Nikber5 (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
...
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+
+        Node(T value, Node<T> next) {
+            this.value = value;
+            this.next = next;
+        }
+
+        Node(T value) {
```
RomanDubovskyi, line 18: we don't use this constructor, let's remove it for now

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
...
+        checkForHead();
+        return head.value;
     }
 
     public T pop() {
-        return null;
+        checkForHead();
+        size--;
+        T deletedValue = head.value;
+        if (head.next != null) {
```
RomanDubovskyi, line 37: it looks like we don't need this if, if it's null you will just set head to null but you check in your checkForHead anyway

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #95, opened 2021-04-16
Added Stack implementation

SakVladyslav:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int MAX_SIZE = 100;
+    private int top;
```
KseniiaMakarova, line 7:
```suggestion
    private int size;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
+    private T[] stack = (T[]) new Object[MAX_SIZE];
```
KseniiaMakarova, line 8: let's move the initialization to the constructor

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
 
+    public MateStack() {
+        top = 0;
+    }
+
+    public void push(T value) {
+        if (top >= MAX_SIZE - 1) {
+            throw new EmptyStackException();
```
KseniiaMakarova, line 16: is that logically ok to throw EmptyStackException here?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
+        } else {
+            top++;
+            stack[top] = value;
```
KseniiaMakarova, line 19: the first element will have index 1 in your array, why so?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
+    private T[] stack = (T[]) new Object[MAX_SIZE];
 
+    public MateStack() {
+        top = 0;
+    }
+
+    public void push(T value) {
+        if (top >= MAX_SIZE - 1) {
+            throw new EmptyStackException();
+        } else {
```
KseniiaMakarova, line 17: do we need else after throw?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
+            stack[top] = value;
+        }
     }
 
     public T peek() {
-        return null;
+        if (top <= 0) {
+            throw new EmptyStackException();
+        } else {
+            return stack[top];
```
KseniiaMakarova, line 27:
```suggestion
            return stack[size - 1];
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
+            top++;
+            stack[top] = value;
+        }
     }
 
     public T peek() {
-        return null;
+        if (top <= 0) {
+            throw new EmptyStackException();
+        } else {
```
KseniiaMakarova, line 26: do we need else after throw?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
+            return stack[top];
+        }
     }
 
     public T pop() {
-        return null;
+        if (top <= 0) {
+            throw new EmptyStackException();
+        } else {
```
KseniiaMakarova, line 34: same

KseniiaMakarova (changes requested):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,45 @@
...
+    private T[] stack;
+
+    public MateStack() {
+        stack = (T[]) new Object[MAX_SIZE];
+        size = 0;
+    }
+
     public void push(T value) {
+        if (size >= MAX_SIZE - 1) {
+            throw new StackOverflowError();
```
KseniiaMakarova, line 17: let's throw a RuntimeException with an informative message

we should not throw Errors

KseniiaMakarova (changes requested):

src/main/java/core/stack/MateStack.java
SakVladyslav: Why do we need to throw  RunTimeException if java already has an error that according to this situation? (It's only for my information)

KseniiaMakarova (approved):

PR #96, opened 2021-04-17
added first solution

IhorRome:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        final T element = data[size - 1];
+        System.arraycopy(data, 0, data, 0, size - 1);
+        size--;
+        return element;
```
andriiromaniuk, line 38: You don't need to copy array each time you delete element, work with what you have.

andriiromaniuk (changes requested):

andriiromaniuk (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+    public void push(T value) {
+        if (size == data.length) {
+            resize();
+        }
+        data[size++] = value;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
```
RomanDubovskyi, line 24: you can move this check to separate method

RomanDubovskyi (approved):

PR #97, opened 2021-04-18
first solution

andriiromaniuk:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private int size;
+    private Node<T> topNode;
+
+    public MateStack() {
+        size = 0;
+    }
```
IhorRome, line 11: you dont have to initialize this variable - it equals 0 by default

IhorRome (changes requested):

IhorRome (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+            Node<T> newNode = new Node<>(value);
+            topNode = newNode;
+            size++;
+            return;
+        }
+        Node<T> newNode = new Node<>(value);
+        newNode.prevNode = topNode;
+        topNode = newNode;
+        size++;
     }
```
LayRipper, line 29:
```suggestion
       Node<T> newNode = new Node<>(value);
        if (topNode != null) {
            newNode.prevNode = topNode; 
        }
        topNode = newNode;
        size++;
    }
```

LayRipper (changes requested):

src/main/java/core/stack/MateStack.java
andriiromaniuk: thanks!

LayRipper (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
...
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        return topNode.value;
     }
 
     public T pop() {
-        return null;
+        if (topNode == null) {
```
RomanDubovskyi, line 35: use same comparison as in peek, and move this code to method since it repeats

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
...
+        private T value;
+        private Node<T> prevNode;
 
+        public Node(T value) {
+            this.value = value;
+        }
+    }
+
+    public void push(T value) {
+        Node<T> newNode = new Node<>(value);
```
RomanDubovskyi, line 19: can we set prev in constructor, so we don't need the following if block

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #99, opened 2021-04-19
added solution

dvdovynets:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,54 @@
...
+        }
+    }
+
+    private static class Node<T> {
+        private final T value;
+        private Node<T> previous;
+
+        public Node(T value, Node<T> previous) {
+            this.value = value;
+            this.previous = previous;
```
freedmanMain, line 51:
```suggestion
        private final T value;
        private Node<T> next;

        public Node(T value, Node<T> next) {
            this.value = value;
            this.next = next;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,54 @@
...
     }
 
     public T pop() {
-        return null;
+        checkStackSize();
+        Node<T> poppedNode = tail;
+        tail = tail.previous;
+        poppedNode.previous = null;
+        size--;
+        return poppedNode.value;
```
freedmanMain, line 32: Same.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,54 @@
...
 
+    public void push(T value) {
+        Node<T> node = new Node<>(value, null);
+        if (head == null) {
+            head = node;
+        } else {
+            node.previous = tail;
+        }
+        tail = node;
+        size++;
```
freedmanMain, line 18: If you take the previous point into account, this method can be written in a couple of lines of code.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,54 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
+    private Node<T> head;
+    private int size;
```
freedmanMain, line 8: You don't need two pointers in a singly-connected list. Remember how you added elements to the head of a singly linked list. Create a node that always points to the top element of the stack. You have to push the elements down.
```suggestion
    private Node<T> top;
    private int size;
```

freedmanMain (changes requested):
https://computersciencewiki.org/index.php/Stack

src/main/java/core/stack/MateStack.java
dvdovynets: Changed but left Previous

src/main/java/core/stack/MateStack.java
dvdovynets: done

src/main/java/core/stack/MateStack.java
dvdovynets: done

src/main/java/core/stack/MateStack.java
dvdovynets: changed

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,41 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private static class Node<T> {
+        private final T value;
+        private final Node<T> previous;
```
freedmanMain, line 34:
```suggestion
        private final Node<T> next;
```

freedmanMain (approved):

src/main/java/core/stack/MateStack.java
dvdovynets: For me more logical is previous. I think it is not matter much

Sofasmile (approved):

PR #100, opened 2021-04-19
added solution

safordog:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,62 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+import java.util.Objects;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int INITIAL_ARRAY_SIZE = 10;
+    private T[] stack;
+    private int capacity;
```
KseniaObertii, line 9:
```suggestion
    private int size;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,62 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+import java.util.Objects;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int INITIAL_ARRAY_SIZE = 10;
```
KseniaObertii, line 7:
```suggestion
    private static final int INITIAL_CAPASITY = 10;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,62 @@
...
 
     public T peek() {
-        return null;
+        checkStackForEmptiness();
+        return stack[capacity - 1];
     }
 
     public T pop() {
-        return null;
+        checkStackForEmptiness();
```
KseniaObertii, line 28: u have to make the element(that u pop) null in the array

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,62 @@
...
+        return stack[--capacity];
     }
 
     public int size() {
-        return 0;
+        return capacity;
+    }
+
+    private boolean contains(T value) {
```
KseniaObertii, line 36: If I were u I would join methods contains() and push(), u used it only once
But it's okay, I think

KseniaObertii (changes requested):

src/main/java/core/stack/MateStack.java
safordog: Дякую за перевірку!!!

KseniaObertii (changes requested):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,64 @@
...
+        for (int i = 0; i <= size; i++) {
+            if (Objects.equals(stack[i], value)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private T[] resizeStack() {
```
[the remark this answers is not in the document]
safordog: Дякую!

KseniaObertii (approved):
Good job))

safordog:
Thanks a lot!

пн, 19 квіт. 2021, 22:37 користувач KseniaObertii ***@***.***>
пише:

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,63 @@
...
+    private T[] stack;
+    private int size;
 
+    public MateStack() {
+        stack = (T[]) new Object[INITIAL_CAPACITY];
+    }
+
+    public void push(T value) {
+        checkCapacity();
+        if (!contains(value)) {
```
RomanDubovskyi, line 17: why do we need to check for contains, we can store to similiar objects

RomanDubovskyi (changes requested):
we don't need contains method, rest is good

RomanDubovskyi (approved):

PR #103, opened 2021-04-19
add solution

KononenkoKonstantin:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+    private static final int START_ARRAY_CAPACITY = 10;
+    private static final double CAPACITY_ARRAY_MULTIPLE = 1.5;
+    private int size;
+    private T[] stackArray;
+    private int capacity;
 
+    public MateStack() {
+        capacity = START_ARRAY_CAPACITY;
+        stackArray = (T[]) new Object[capacity];
     }
```
Roman0607, line 15: try to rewrite this task using class Node

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+        size++;
     }
 
-    public T pop() {
-        return null;
+    public T peek() throws EmptyStackException {
+        if (size > 0) {
+            return stackArray[size - 1];
+        }
+        throw new EmptyStackException();
```
Roman0607, line 27: create separate method for checking size

Roman0607 (changes requested):

src/main/java/core/stack/MateStack.java
Roman0607: It also good solution, but in my opinion when we use class Node it is easier. As you wish

KononenkoKonstantin: done

src/main/java/core/stack/MateStack.java
KononenkoKonstantin: done

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+    private int size;
 
+    public void push(T value) {
+        Node<T> node;
+        if (size == 0) {
+            node = new Node<>(value, null);
+        } else {
+            node = new Node<>(value, top);
+        }
+        top = node;
```
Max-Pochepets, line 16:
```suggestion
        top = new Node<>(value, top);
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+        size++;
     }
 
-    public T peek() {
-        return null;
+    public T peek() throws EmptyStackException {
```
Max-Pochepets, line 20:
```suggestion
    public T peek() {
```
is it checked or unchecked?

Max-Pochepets (changes requested):

src/main/java/core/stack/MateStack.java
KononenkoKonstantin: it's unchecked

src/main/java/core/stack/MateStack.java
KononenkoKonstantin: done

src/main/java/core/stack/MateStack.java
Max-Pochepets: than we don't need to even process it with try-catch, not to mention throwing it up

src/main/java/core/stack/MateStack.java
KononenkoKonstantin: I've removed block if - else here, because is size ==0, top ==null

Max-Pochepets (approved):

PR #104, opened 2021-04-19
"added solution"

KseniaObertii:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+        }
+        return top.value;
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        Node<T> requiredNode = top;
```
freedmanMain, line 39: you can simply call the Peek() method here.
```suggestion
        Node<T> requiredNode = Peek();
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+            this.previous = previous;
+        }
+    }
+
+    public void push(T value) {
+        if (size == 0) {
+            top = new Node<>(value, null);
+        } else {
+            top = new Node<>(value, top);
+        }
```
freedmanMain, line 24: Initially when the stack is empty the Top field is null. you don't need this check here.
```suggestion
            top = new Node<>(value, top);
```

freedmanMain (changes requested):

src/main/java/core/stack/MateStack.java
KseniaObertii: done

freedmanMain (approved):
xorow!

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,41 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> top;
+    private int size;
+
+    private static class Node<T> {
```
RomanDubovskyi, line 9: it's better to move private class to the end of the file

RomanDubovskyi (approved):
good job

RomanDubovskyi (approved):

PR #109, opened 2021-04-20
Initial implementation of Stack

Den4eg007:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
```
KseniaObertii, line 6:
```suggestion
    private Node<T> top;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+    private int size;
 
+    public void push(T value) {
+        if (tail == null) {
+            tail = new Node<>(value, null);
+        } else {
+            tail = new Node<>(value, tail);
+        }
```
KseniaObertii, line 14: top is already null, we don't need first check
```suggestion
            tail = new Node<>(value, tail);
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
     public T peek() {
-        return null;
+        checkSize();
+        return tail.value;
     }
 
     public T pop() {
-        return null;
+        checkSize();
+        T element = tail == null ? null : tail.value;
```
KseniaObertii, line 25:
```suggestion
        T element = peek()
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+        tail = tail == null ? null : tail.prev;
```
KseniaObertii, line 26: we don't need this check
```suggestion
        tail = tail.prev;
```

KseniaObertii (changes requested):

KseniaObertii (approved):
That's it, man

Max-Pochepets (approved):

PR #113, opened 2021-04-20
added solution

iedenysenko:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int ARRAY_SIZE = 10;
```
Den4eg007, line 6: maybe rename it to INITIAL_ARRAY_SIZE

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
...
+    private static final int FIRST_INDEX = 0;
+    private static final int DECREMENT = 1;
```
Den4eg007, line 8: I think that these would be self-explanatory written directly in code, without creating constants

Den4eg007 (approved):

src/main/java/core/stack/MateStack.java
Max-Pochepets: ++

src/main/java/core/stack/MateStack.java
Max-Pochepets: ++

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
...
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        return array[size - DECREMENT];
     }
 
     public T pop() {
-        return null;
+        T removed;
+        removed = peek();
```
Max-Pochepets, line 33:
```suggestion
        T removed = peek();
```

Max-Pochepets (changes requested):

src/main/java/core/stack/MateStack.java
iedenysenko: done

src/main/java/core/stack/MateStack.java
iedenysenko: done

src/main/java/core/stack/MateStack.java
iedenysenko: done

Max-Pochepets (approved):

PR #114, opened 2021-04-20
Task was completed

divertitoS:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
-        return 0;
+        return size;
+    }
+
+    private void resize() {
+        if (array.length == size) {
+            T[] newArray = (T[]) new Object[DEFAULT_CAPACITY * 3 / 2];
+            System.arraycopy(array, 0, newArray, 0, size);
+            array = newArray;
+        }
```
Vlad-Andr, line 43: I did the same and mentor told me that it should be two methods (check and grow).

Vlad-Andr (approved):

src/main/java/core/stack/MateStack.java
divertitoS: Thanks

Max-Pochepets (approved):

PR #117, opened 2021-04-20
Finish homework requirements.

dant0n:


Nadhob (approved):
Everything is ok.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,45 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int DEFAULT_STACK_SIZE = 10;
+    private T[] stackArray = (T[]) new Object[DEFAULT_STACK_SIZE];
```
Max-Pochepets, line 7:
```suggestion
    private T[] stackArray;
```
there is a constructor for this

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,45 @@
...
+
+    private void checkSize() {
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    private void resize() {
+        int newSize = stackArray.length * 3 / 2 + 1;
+        T[] newArray = (T[]) new Object[newSize];
```
Max-Pochepets, line 41:
```suggestion
        T[] newArray = (T[]) new Object[stackArray.length * 3 / 2 + 1];
```
why +1?

Max-Pochepets (changes requested):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+    }
+
+    private void checkSize() {
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    private void resize() {
+        T[] newArray = (T[]) new Object[stackArray.length * 3 / 2 + 1];
```
Max-Pochepets, line 44: and do we even need to grow the array when size is 0 or 1, considering default capacity is 10 and we call this method when `capacity == size`?

Max-Pochepets (changes requested):

src/main/java/core/stack/MateStack.java
dant0n: Yes, because we can change the default capacity to 0 or 1. Or implement a trim method that will reduce the size of the array after each pop() call.

Max-Pochepets (approved):
👍

PR #119, opened 2021-04-20
Basic Stack implemented.

Nadhob:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int DEFAULT_STACK_SIZE = 10;
+    private static final int GROW_MULTIPLIER = 2;
+    private Object[] stack;
+    private int size;
+
+    public MateStack() {
+        stack = new Object[DEFAULT_STACK_SIZE];
+        size = 0;
```
dant0n, line 13: No need to initialise size here, it has a default value of 0;

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+    }
 
+    public void push(T value) {
+        if (stack.length >= size) {
+            grow();
+        }
```
dant0n, line 19: stack.length will almost always be larger than size
```suggestion
        if (size >= stack.length) {
            grow();
        }
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+        return (T) stack[size - 1];
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        T lastElement = (T) stack[size - 1];
+        stack[size-- - 1] = null;
```
dant0n, line 35:
```suggestion
        stack[--size] = null;
```

dant0n (changes requested):
Few comments:)

dant0n (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+            grow();
+        }
+        stack[size++] = value;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
Max-Pochepets, line 25: this code repeats twice

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #120, opened 2021-04-20
hw-stack

Trident1998:
There was implemented the class MateStack

### DIFF
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -3,19 +3,12 @@ package core.stack;
 import java.util.EmptyStackException;
 
 public class MateStack<T> {
-    private Node<T> head;
     private Node<T> tail;
     private int size;
 
     public void push(T value) {
-        Node<T> newElement = new Node<>(value, null);
-        if (size == 0) {
-            tail = newElement;
-            head = newElement;
-        } else {
-            tail.next = newElement;
-            tail = newElement;
-        }
+        Node<T> newElement = new Node<>(value, tail);
+        tail = newElement;
         size++;
     }
 
@@ -27,8 +20,8 @@ public class MateStack<T> {
     public T pop() {
         checkException();
         Node<T> oldElement = tail;
-        tail = findElement();
-        tail.next = null;
+        tail = tail.next;
+        size--;
         return oldElement.item;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,63 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> head;
```
Kharlanava, line 6: Do we really need Node<T> head here?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,63 @@
...
-        return null;
+        checkException();
+        return tail.item;
     }
 
     public T pop() {
-        return null;
+        checkException();
+        Node<T> oldElement = tail;
+        tail = findElement();
```
Kharlanava, line 30: Isn't it too complex? You update tail each time you push new value

Kharlanava (changes requested):

src/main/java/core/stack/MateStack.java
Trident1998: no its variable is redundant!

src/main/java/core/stack/MateStack.java
Trident1998: I fixed this

Kharlanava (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail;
```
Max-Pochepets, line 6:
```suggestion
    private Node<T> top;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,47 @@
...
+    private int size;
 
+    public void push(T value) {
+        Node<T> newElement = new Node<>(value, tail);
+        tail = newElement;
```
Max-Pochepets, line 11:
```suggestion
        tail = new Node<>(value, tail);
```

Max-Pochepets (approved):

PR #125, opened 2021-04-20
completed task

mmelniichuk:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+            grow();
+        }
+        elementData[size++] = value;
     }
 
     public T peek() {
-        return null;
+        if (empty() == true) {
+            throw new EmptyStackException();
+        }
```
Den4eg007, line 24: You can move this to empty() method and use it here

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+        return elementData[size - 1];
     }
 
     public T pop() {
-        return null;
+        if (empty() == true) {
+            throw new EmptyStackException();
+        }
```
Den4eg007, line 31: same

Den4eg007 (changes requested):

src/main/java/core/stack/MateStack.java
mankokolya: Call this only in peek as you call peek on line 32 and use only empty() without == true as empty() returns boolean

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+
+    public boolean empty() {
+        return size == 0;
+    }
+
+    private void grow() {
+        T[] newElementData = (T[]) new Object[elementData.length
+                                        + elementData.length / 2];
+        System.arraycopy(elementData, 0, newElementData, 0, size);
+        elementData = newElementData;
```
mankokolya, line 48: you can do this in one line Arrays.copyOf(elementData, new size)

mankokolya: elementData = Arrays.copyOf(elementData, new size)

mankokolya (changes requested):

src/main/java/core/stack/MateStack.java
mmelniichuk: done! good idea

src/main/java/core/stack/MateStack.java
mmelniichuk: done

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+        elementData[size--] = null;
+        return removedElement;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    public boolean empty() {
```
Den4eg007, line 37: You can make this method void

mmelniichuk: I tried to override default empty() method, just like the rest) so it must be public boolean...

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
 
     public T peek() {
-        return null;
+        empty();
+        return elementData[size - 1];
     }
 
     public T pop() {
-        return null;
+        empty();
```
mankokolya, line 27: as you are calling peek() on line 28 you don't need this empty(). And you can remove empty at all jest check size in peek and throw exception

mankokolya (changes requested):

src/main/java/core/stack/MateStack.java
mmelniichuk: done

src/main/java/core/stack/MateStack.java
mmelniichuk: super! thank's

Den4eg007 (approved):

RomanDubovskyi (approved):

PR #133, opened 2021-04-20
C001 - All done! For review!

Koziarevskyi:


ihorromaniuk (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> last;
```
Max-Pochepets, line 6:
```suggestion
    private Node<T> top;
```
it's not a list. consider this as a tower of some sort

Max-Pochepets (approved):

PR #134, opened 2021-04-20
added solution

SerKisJD:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
+    public void push(T value) {
+        current = new Node<>(current, value);
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (current == null) {
+            throw new EmptyStackException();
+        }
```
Roman0607, line 17: Good job, actually I think you can create separate method for this check, because you used it several times)

Roman0607 (changes requested):

Roman0607 (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private int size;
+    private Node<T> current;
```
Max-Pochepets, line 7:
```suggestion
    private Node<T> top;
```

Max-Pochepets (approved):

PR #136, opened 2021-04-21
added solution

PepsiTheDoge:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
     public T peek() {
-        return null;
+        checkSize();
+        return head.value;
     }
 
     public T pop() {
-        return null;
+        checkSize();
+        T poppedValue = head.value;
```
Den4eg007, line 31:
```suggestion
        T poppedValue = peek();
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private int size;
+    private Node<T> head;
+
+    private static class Node<T> {
```
Den4eg007, line 9: maybe move this to the bottom of the class

PepsiTheDoge: but I like it here...

darinaTo (approved):

src/main/java/core/stack/MateStack.java
RomanDubovskyi: usually private elements are moved to the bottom of the class but it's not crucial

RomanDubovskyi (approved):

PR #137, opened 2021-04-21
HW Stack

veronika1shcherbyna:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
     public T peek() {
-        return null;
+        checkNull();
+        return elements[size];
     }
 
     public T pop() {
-        return null;
+        checkNull();
+        return elements[size--];
```
Anatolii-Zarytskyi, line 28: did the value became null after peek? or we still have it in stack?

Anatolii-Zarytskyi (changes requested):

Anatolii-Zarytskyi (approved):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
 
+    public MateStack() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    public void push(T value) {
+        if (size == elements.length) {
+            increaseCapacity();
+        }
+        elements[++size] = value;
```
RomanDubovskyi, line 18:
```suggestion
        elements[size++] = value;
```
when size is 0 you should put an element on the 0 index not 1

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
-        return null;
+        checkNull();
+        return elements[size];
     }
 
     public T pop() {
-        return null;
+        checkNull();
+        T current = peek();
+        elements[size--] = null;
```
RomanDubovskyi, line 29:
```suggestion
        elements[--size] = null;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+        return current;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private void checkNull() {
```
RomanDubovskyi, line 37:
```suggestion
    private void isEmpty() {
```

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #148, opened 2024-04-05
First Attempt

jszymanski97:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int DEF_CAPACITY = 10;
+    private Object[] array;
```
okuzan, line 7: use your class type parameter, that's exactly its purpose

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+    private int size;
 
+    public MateStack() {
+        array = new Object[DEF_CAPACITY];
+        size = 0;
```
okuzan, line 12: 0 is default value for integer

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+        return size;
+    }
+
+    public boolean isEmpty() {
+        return size == 0;
+    }
+
+    public void checkCapacity() {
+        if (size == array.length) {
+            int newSize = array.length * 2;
```
okuzan, line 46: extract 2 to a GROW_FACTOR constant

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
     public int size() {
-        return 0;
+        return size;
+    }
+
+    public boolean isEmpty() {
+        return size == 0;
+    }
+
+    public void checkCapacity() {
```
okuzan, line 44: should these two be public?

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,52 @@
...
+            throw new EmptyStackException();
+        }
+        return (T) array[size - 1];
     }
 
     public T pop() {
-        return null;
+        if (isEmpty()) {
+            throw new EmptyStackException();
+        }
```
okuzan, line 30: better extract to a separate method not only condition checking, but exception throwing too

okuzan (changes requested):

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,54 @@
...
 
+    public void push(T value) {
+        checkCapacity();
+        array[size++] = value;
     }
 
     public T peek() {
-        return null;
+        checkNotEmpty();
+        return array[size - 1];
```
d1sam, line 22: try to avoid magic numbers.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,54 @@
...
+        return current;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private boolean isEmpty() {
+        return size == 0;
```
d1sam, line 37: try to avoid magic numbers.

d1sam (approved):
Well done!

PR #151, opened 2024-06-06
solution

mgsorigo:


src/main/java/core/stack/Node.java
MateuszRostek, line 1: Node should be an inner class in `MateStack`

MateuszRostek: Also you don't need getters/setters to use your inner class then.

MateuszRostek (changes requested):
Good job! See minor advice.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+
+    private static class Node<T> {
+        private final T data;
+        private Node<T> next;
 
+        Node(T data) {
+            this.data = data;
+            this.next = null;
+        }
+    }
```
MateuszRostek, line 22: This should be moved to the end of the `MateStack` class.

MateuszRostek (approved):
Good job!

PR #152, opened 2024-06-13
initial solution

MarcusPy:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,57 @@
...
+    private static int GROWTH_FACTOR;
+    private T[] stack;
+    private int size;
+    private int capacity;
 
+    public MateStack() {
+        GROWTH_FACTOR = 2;
+        capacity = 10;
+        stack = (T[]) new Object[capacity];
+        size = 0;
```
MateuszRostek, line 15: It is redundant, size is 0 by default

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,57 @@
...
+        stack = (T[]) new Object[defaultCapacity];
+        size = 0;
+    }
+
+    public void push(T value) {
+        if (size >= capacity) {
+            grow();
+        }
+        stack[size] = value;
+        size++;
```
MateuszRostek, line 30: image

```suggestion
        stack[size++] = value;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,57 @@
...
+        capacity = 10;
+        stack = (T[]) new Object[capacity];
+        size = 0;
+    }
+
+    public MateStack(int defaultCapacity, int growthFactor) {
+        GROWTH_FACTOR = growthFactor;
+        capacity = defaultCapacity;
+        stack = (T[]) new Object[defaultCapacity];
+        size = 0;
```
MateuszRostek, line 22: It is redundant, size is 0 by default

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,57 @@
...
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private void grow() {
+        T[] oldStack = stack;
+        capacity *= GROWTH_FACTOR;
+        stack = (T[]) new Object[capacity];
+        System.arraycopy(oldStack, 0, stack, 0, size);
```
MateuszRostek, line 55: This is valid approach, but you can also do it like this:

```java
        capacity = (int) (currentSize * DEFAULT_MULTIPLAYER);
        System.arraycopy(stack, 0, stack, 0, capacity);
```

MateuszRostek (approved):
Good job, approved! See minor advice.

PR #153, opened 2024-07-16
My Stack implementation

smagles:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+    private int size;
 
+    public void push(T value) {
+        top = new Node<>(top, value);
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (isEmpty()) {
```
beatarjn, line 15: I would suggest more descriptive name e.g. isStackEmpty()

beatarjn (changes requested):
Well done, small adjustment suggested :)

PR #155, opened 2024-08-09
First solution, basic tasks solved

noiseintheattic:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,57 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private T[] grow() {
+        double newLength = table.length * 1.5;
+        currentSize = (int) newLength;
```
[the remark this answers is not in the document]
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,57 @@
...
+        T[] newTable = (T[]) new Object[currentSize];
+        for (int i = 0; i < table.length; i++) {
+            newTable[i] = table[i];
+        }
```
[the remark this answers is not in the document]
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,57 @@
...
-    public void push(T value) {
+    private static final int INITIAL_SIZE = 10;
+    private int currentSize = INITIAL_SIZE;
+    private int size = 0;
+    private T[] table = (T[]) new Object[currentSize];
 
+    public void push(T value) {
+        if (size >= currentSize) {
+            table = grow();
+        }
```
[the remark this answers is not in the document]
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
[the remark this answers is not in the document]
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,57 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int INITIAL_SIZE = 10;
+    private int currentSize = INITIAL_SIZE;
+    private int size = 0;
+    private T[] table = (T[]) new Object[currentSize];
```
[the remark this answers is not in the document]
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,61 @@
...
-    public void push(T value) {
+    private static final int INITIAL_SIZE = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+    private int currentSize = INITIAL_SIZE;
+    private int size = 0;
+    private T[] table = createArray(currentSize);
 
+    private T[] createArray(int size) {
+        return (T[]) new Object[size];
+    }
```
MateuszRostek, line 14: It would be better to just use constructor here:

```java
    @SuppressWarnings("unchecked")
    public MateStack() {
        this.stack = (T[]) new Object[DEFAULT_SIZE];
        currentSize = INITIAL_SIZE;
    }
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,61 @@
...
+        size--;
+        return value;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private T[] grow() {
```
MateuszRostek, line 43: This name does not describe what this method do. In addition, there is no need to return the table from this method, you can make the necessary changes inside it, and then simply refer to this table in another method.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,61 @@
...
+        if (size < currentSize) {
+            return table;
+        }
+        int newLength = (int) (table.length * GROWTH_FACTOR);
+        currentSize = (int) newLength;
+        T[] newTable = createArray(currentSize);
+        System.arraycopy(table,0,newTable,0,table.length);
+        return newTable;
+    }
```
MateuszRostek, line 52: This could also be simplified like this for example (just do necessary checks in the `push` method before calling this one:

```java
        private void resize() {
    }
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,61 @@
...
+        table = grow();
+        table[size] = value;
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
MateuszRostek, line 25: You are already calling the exact same check 2 times in such a small class - move this logic to a separate private method.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,61 @@
...
+        return table[size - 1];
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
```
MateuszRostek, line 32: You are already calling the exact same check 2 times in such a small class - move this logic to a separate private method

MateuszRostek (changes requested):
Nice try! See some advice and implement it

src/main/java/core/stack/MateStack.java
noiseintheattic: I'm sorry, I don't understand what exactly should I do here...

src/main/java/core/stack/MateStack.java
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
beatarjn: It would be better practice to initialize  the table array not in the method createArray() but in constructor because:
Initialization Control: By initializing the array in the constructor, you have better control over the initialization process and can ensure that the array is always initialized properly when an instance of MateStack is created.

Avoiding Unnecessary Method: Initializing the array directly in the constructor avoids the need for a separate createArray method, which simplifies the code and makes it more straightforward.

Flexibility: Initializing the array in the constructor allows for more flexibility if you want to pass in different sizes or configurations when creating an instance of MateStack.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private void resize() {
+        currentSize = (int) (table.length * GROWTH_FACTOR);
+        System.arraycopy(table,0,table,0,currentSize);
```
beatarjn, line 43: Pls make sure that your code is properly formatted (Ctrl+Alt+L (Windows/Linux) or⌥⌘L (macOS)).
Also it would be better to allocate a new array with the increased size, and then copy the existing elements into the new array.
`currentSize = (int) (table.length * GROWTH_FACTOR);
    T[] newTable = createArray(currentSize);
    System.arraycopy(table, 0, newTable, 0, size);
    table = newTable;`

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
+        return size == 0;
+    }
+
+    private void checkSize() {
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
     }
 }
+
```
beatarjn, line 56: Please remove empty lines

beatarjn (changes requested):

src/main/java/core/stack/MateStack.java
noiseintheattic: ok, done

src/main/java/core/stack/MateStack.java
noiseintheattic: ok, done

MateuszRostek (approved):
Good job!

PR #156, opened 2024-08-30
commit task

Oleksii21th:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,65 @@
...
+
     public void push(T value) {
+        if (value == null) {
+            return;
+        }
+
+        resize();
+        elementData[size] = value;
+        size++;
+    }
```
MateuszRostek, line 24: why do always call resize() method? you should only resize if `size + 1 == currentMaxSize`. 
Consider adding: 
`size` variable to keep track of the current size
`currentMaxSize` variable to keep track of currently maximum size.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,65 @@
...
+            System.arraycopy(elementData, 0, newArray, 0, elementData.length);
+            elementData = newArray;
+        }
     }
 
     public T peek() {
-        return null;
+        if (isEmpty()) {
+            throw new EmptyStackException();
+        }
```
MateuszRostek, line 42: This code is repeated in peek and pop, lets create separate method for it, for example like this (then you can also remove the `isEmpty()` method):

```java
private void checkIfStackEmpty() {
        if (size == 0) {
            throw new EmptyStackException();
        }
    }
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,65 @@
...
+            int newCapacity = elementData.length * 2;
+            if (newCapacity - MAX_ARRAY_SIZE > 0) {
+                newCapacity = MAX_ARRAY_SIZE;
+            }
 
+            T[] newArray = (T[]) new Object[newCapacity];
+            System.arraycopy(elementData, 0, newArray, 0, elementData.length);
+            elementData = newArray;
+        }
     }
```
MateuszRostek, line 37: after changes, this method can be simplified to something like this (Creating constant for default multiplayer is good approach in this case):

```java
private void resize() {
    }
```

MateuszRostek (changes requested):
Good job! See some advice and implement required changes:)

MateuszRostek (approved):
Good job!

PR #158, opened 2024-10-17
Bugged solution of the task

B4rt1omiej:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
+
 package core.stack;
 
+import java.util.Arrays;
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int DEFAULT_SIZE = 8;
+    private Object[] stack;
```
slade13, line 9: The stack array should be of type T instead of Object for type safety. Replace `private Object[] stack;` with `private T[] stack;` and adjust the rest of your code accordingly.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+    private int size = 0;
 
+    public MateStack() {
+        stack = new Object[DEFAULT_SIZE];
+        Arrays.fill(stack, null);
+    }
+
+    public void push(T value) {
+        if (size < stack.length) {
```
slade13, line 18: The stack implementation should not have a fixed size. Instead of throwing StackOverflowError when the array is full, you should create a larger array and copy the old elements into it.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+            return (T) stack[size - 1];
+        } else {
+            throw new EmptyStackException();
+        }
     }
 
     public T pop() {
-        return null;
+        if (size > 0) {
+            Object value = stack[size - 1];
```
slade13, line 36: It's unnecessary to set the value to null before reducing the size. JVM garbage collector will clean up the unused memory automatically.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,49 @@
...
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int DEFAULT_SIZE = 8;
+    private Object[] stack;
+    private int size = 0;
 
+    public MateStack() {
+        stack = new Object[DEFAULT_SIZE];
+        Arrays.fill(stack, null);
```
slade13, line 14: There is no need to fill the stack array with nulls during initialization. New arrays in Java are already filled with nulls.

slade13 (changes requested):
Please update your code based on my comments.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+    private final int multiplier = 2;
+
+    public MateStack() {
+        stack = new Object[DEFAULT_SIZE];
+    }
 
+    public void grow() {
+        if (size < stack.length) {
+            int newCapacity = size * multiplier;
+            System.arraycopy(stack, 0, stack = new Object[newCapacity], 0, size);
```
MateuszRostek, line 20:
```suggestion
            System.arraycopy(stack, 0, stack, 0, size);
```

MateuszRostek (changes requested):
If you have troubles implementing comments added by mentor, please try to find solution using internet/AI and if you can't find it, ask for help on chat instead of re-requesting review without fulfilling requirements.

src/main/java/core/stack/MateStack.java
B4rt1omiej: why like this? if i do copy of exacly same array without growing its size then method grow is copletly usless. Also what about previeus comment? shall i change Object[] to T[] or not?

PR #159, opened 2024-10-20
first version

MateuszSeler:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+        } else {
+            Element<T> notTheTopAnyMore = top;
+            top = new Element<>(value, notTheTopAnyMore);
+        }
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
```
slade13, line 30: In the 'peek' method, to make your code cleaner, you can avoid using 'else' after 'throw'. When the 'throw' statement is executed, the control is immediately transferred out of the method. Therefore, there is no need to use 'else'.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+            throw new EmptyStackException();
+        } else {
+            return top.value;
+        }
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
```
slade13, line 39: The same comment as above applies to the 'pop' method. You can avoid using 'else' after 'throw'.

slade13 (approved):
Good job! Please review my comments.

PR #160, opened 2024-10-23
My first approach to the stack task

Pawel-Cp:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
...
 public class MateStack<T> {
+    private Element<T> topOfStack;
+    private int size;
+
+    private static class Element<T> {
+        private Element<T> prev;
+        private T value;
+
+        public Element(Element<T> prev, T value) {
+            this.prev = prev;
```
slade13, line 14: The constructor of the nested Element class is public. It should be private to ensure encapsulation as it is used only within the MateStack class.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,53 @@
...
+
     }
 
     public T pop() {
-        return null;
+        sizeCheck();
+        Element<T> oldTopValue = topOfStack;
+        topOfStack = oldTopValue.prev;
+        size--;
+        return oldTopValue.value;
```
slade13, line 41: It would be better to nullify the 'prev' of the popped element to help garbage collector.

slade13 (approved):
Good job!

PR #162, opened 2024-11-13
implement methods in class MateStack

vShmandiuk:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,70 @@
...
+        }
+
+        elementData[size] = value;
+        size++;
+    }
+
+    private void resize() {
+        if (size == elementData.length) {
+            int newCapacity = currentMaxSize * DEFAULT_MULTIPLIER;
+            if (newCapacity - MAX_ARRAY_SIZE > 0) {
```
slade13, line 34: The expression 'newCapacity - MAX_ARRAY_SIZE > 0' is always false, because the maximum possible newCapacity can be MAX_ARRAY_SIZE. You should remove this redundant check.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,70 @@
...
+    private int size;
+    private int currentMaxSize;
+
+    public MateStack() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+        currentMaxSize = DEFAULT_CAPACITY;
+    }
+
     public void push(T value) {
+        if (value == null) {
```
slade13, line 20: The method push should not ignore null values. If the user wants to add a null value to the stack, it should be allowed.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,70 @@
...
+
+    private void resize() {
+        if (size == elementData.length) {
+            int newCapacity = currentMaxSize * DEFAULT_MULTIPLIER;
+            if (newCapacity - MAX_ARRAY_SIZE > 0) {
+                newCapacity = MAX_ARRAY_SIZE;
+            }
 
+            T[] newArray = (T[]) new Object[newCapacity];
+            System.arraycopy(elementData, 0, newArray, 0, elementData.length);
```
slade13, line 39: Array copying and reference swapping should be atomic, to avoid data corruption in case of concurrent access. Consider using 'Arrays.copyOf' method, which handles this automatically.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,70 @@
...
+        if (size == currentMaxSize) {
+            resize();
+        }
+
+        elementData[size] = value;
+        size++;
+    }
+
+    private void resize() {
+        if (size == elementData.length) {
```
slade13, line 32: The condition 'size == elementData.length' is always true when resizing is needed, because resize is only called when 'size == currentMaxSize', and 'currentMaxSize' is always equal to 'elementData.length'. You can remove this redundant check.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,70 @@
...
+
+    private boolean isEmpty() {
+        return size == 0;
     }
 
     public T pop() {
-        return null;
+        checkIfStackEmpty();
+        size--;
+        T element = (T) elementData[size];
```
slade13, line 62: The cast to T is not necessary. The array is already of type T[], so the elements are of type T.

slade13 (approved):
Good job with implementing this task. Please review my comments.

PR #163, opened 2025-02-04
Initial commit.

Wojtek-A-JAVA:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+        }
+        T[] tempStack = (T[]) new Object[newCapacity];
+        if (newCapacity > capacity) {
+            System.arraycopy(mateStack, 0, tempStack, 0, capacity);
+        }
+        if (newCapacity < capacity) {
+            System.arraycopy(mateStack, 0, tempStack, 0, capacity - 1);
+        }
+        mateStack = tempStack;
+        tempStack = null;
```
slade13, line 53: Setting `tempStack` to null is unnecessary as it will be garbage collected once it goes out of scope. This line can be removed.
```suggestion
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+    private static final int STACK_ENLARGEMENT = 1;
+    private static final int STACK_DECREMENT = -1;
+    private T[] mateStack;
+
+    public MateStack() {
+        this.mateStack = (T[]) new Object[EMPTY_STACK_SIZE];
+    }
 
+    public void push(T value) {
+        changeStackSize(STACK_ENLARGEMENT);
```
slade13, line 16: The `changeStackSize` method is called with `STACK_ENLARGEMENT`, which increases the array size by 1 each time an element is pushed. This can lead to inefficient memory usage and performance issues. Consider implementing a dynamic resizing strategy, such as doubling the array size when it's full. 
In case of 20 additions, your code will create 20 new arrays.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+
 public class MateStack<T> {
-    public void push(T value) {
+    private static final int EMPTY_STACK_SIZE = 0;
+    private static final int STACK_ENLARGEMENT = 1;
+    private static final int STACK_DECREMENT = -1;
+    private T[] mateStack;
+
+    public MateStack() {
+        this.mateStack = (T[]) new Object[EMPTY_STACK_SIZE];
```
slade13, line 12: Initializing the stack with an array of size 0 is inefficient. Consider starting with a default size, such as 10, to avoid frequent resizing when elements are added - just like the comment from AI mentor says.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+        if (mateStack.length > EMPTY_STACK_SIZE) {
+            return mateStack[mateStack.length - 1];
+        }
+        throw new EmptyStackException();
     }
 
     public T pop() {
-        return null;
+        T value = peek();
+        changeStackSize(STACK_DECREMENT);
```
slade13, line 29: The `changeStackSize` method is called with `STACK_DECREMENT`, which decreases the array size by 1 each time an element is popped. This can lead to inefficient memory usage and performance issues. Consider implementing a dynamic resizing strategy, such as halving the array size when it's less than a quarter full.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,55 @@
...
+
+        if (newCapacity == EMPTY_STACK_SIZE) {
+            mateStack = (T[]) new Object[EMPTY_STACK_SIZE];
+            return;
+        }
+        T[] tempStack = (T[]) new Object[newCapacity];
+        if (newCapacity > capacity) {
+            System.arraycopy(mateStack, 0, tempStack, 0, capacity);
+        }
+        if (newCapacity < capacity) {
```
slade13, line 49: You can replace this opposite condition with else statement.
```suggestion
        } else {
```

slade13 (approved):
Good job, there are some performance problems but nevertheless your code will work as intended to.

PR #166, opened 2025-02-09
firts commit

lisichenkowork:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,69 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private final int initCapacity = 16;
+    private final int growFactor = 2;
+    private T[] elementData;
+    private int size = 0;
```
MateuszRostek, line 9: redundant, it's zero by default

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,69 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private final int initCapacity = 16;
+    private final int growFactor = 2;
```
MateuszRostek, line 7: These should be constants, `private static final int INIT_CAPACITY = ...` for example

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,69 @@
...
+    private int size = 0;
+
+    @SuppressWarnings("unchecked")
+    public MateStack() {
+        this.elementData = (T[]) new Object[initCapacity];
+    }
 
+    public void push(T value) {
+        if (size == elementData.length) {
+            reSize(elementData.length * growFactor);
```
MateuszRostek, line 18: logic of calculating new size should be directly in `resize` method, therefore resize should not take any params

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,69 @@
...
+        return this.size;
+    }
+
+    private void checkIfEmpty() {
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    private void reSize(int newCapacity) {
```
MateuszRostek, line 47:
```suggestion
    private void resize(int newCapacity) {
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,69 @@
...
+        @SuppressWarnings("unchecked")
+        T[] newArr = (T[]) new Object[newCapacity];
+        System.arraycopy(elementData, 0, newArr, 0, size);
+        elementData = newArr;
+    }
```
MateuszRostek, line 52: It can be simplfied to something like this:

```java
    private void resize() {
    }
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,69 @@
...
+
+    @Override
+    public String toString() {
```
MateuszRostek, line 55: You were asked to implement all methods in the class `MateStack`, please do not add extra code, like this method

MateuszRostek (approved):
Good job, see some minor advice

PR #167, opened 2025-02-11
Tasks ready

KamilSzerszen:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
 public class MateStack<T> {
-    public void push(T value) {
 
+    private int size = 0;
+    private int index = 0;
+    private Object[] array = new Object[10];
+
+    public void push(T value) {
+        ensureCapacity();
+        size++;
```
slade13, line 14: The 'size' variable is incremented before adding the element to the array. This can lead to inconsistencies if an exception occurs during the operation. Consider incrementing 'size' after successfully adding the element.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
 package core.stack;
 
+import java.util.Arrays;
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
 
+    private int size = 0;
+    private int index = 0;
```
slade13, line 9: The 'index' variable is redundant since it mirrors the 'size' variable. You can use 'size' to track the current index for the next push operation, which will simplify the code.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private void ensureCapacity() {
+        if (index == array.length) {
+            array = Arrays.copyOf(array, array.length << 1);
```
slade13, line 43: Using 'array.length << 1' is a bitwise operation to double the array size. While this works, it might be less readable for some developers. Consider using 'array.length * 2' for clarity.

slade13 (approved):
Good job, please review my comments.

PR #168, opened 2025-02-21
First solution

KIR4S3K:


src/main/java/core/stack/Main.java
MateuszRostek, line 1: Don't push the main class where you test your solution (unless specified in requirements)

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,67 @@
...
+        newNode.setNext(top);
+        top = newNode;
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (top == null) {
+            throw new EmptyStackException();
+        }
```
MateuszRostek, line 50: This check is used in two places (peek and pop methods) consider extracting it to the separate method to avoid redundant duplication

MateuszRostek (approved):
Good job, approved! See minor advice

PR #171, opened 2025-03-18
solution

weronikarzekiecka:


MateuszRostek (approved):
Good job! 
Please fix this error - that's why checks are failing now:
`Error:  /home/runner/work/jv-stack/jv-stack/src/main/java/core/stack/MateStack.java:1: File does not end with a newline. [NewlineAtEndOfFile]`

PR #173, opened 2025-04-05
implementation of class Stack

zabmari:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (head == null) {
+            throw new EmptyStackException();
+        }
+        T value = head.value;
+        return value;
```
MateuszRostek, line 20: Do not create redundant variables, just return it when it's possible
```suggestion
        return head.value;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,46 @@
...
+        Node<T> node = new Node<T>(value, head);
+        head = node;
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (head == null) {
+            throw new EmptyStackException();
+        }
```
MateuszRostek, line 18: You are using this logic in peek and pop methods, it should be extracted to separate method

MateuszRostek (approved):
Good job!

PR #176, opened 2025-05-26
first solution

anpiet:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,30 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+import java.util.LinkedList;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private LinkedList<T> elements = new LinkedList<>();
```
slade13: Don’t use java.util.LinkedList, in task description there is note to not use existing collections:

> Your implementation should be based on the array or singly linked list (don't use existing collections).

Introduce your `Node` class:
```
private static class Node<T> {
    T value;
    Node<T> next;
    Node(T value, Node<T> next) { this.value = value; this.next = next; }
}
```
Keep a `private Node<T> head;` and an `int size;`

`push `→ link a new node at `head`, increment `size`.

`pop `/ `peek `→ operate on `head`, throw if `head == null`.

slade13 (changes requested):
You are not allowed to use existing Java collections, please check my comment.

anpiet:
implemented new solution according to comments

slade13 (approved):
Good job!

PR #180, opened 2025-09-15
Implementation of stack

nowakjd:
Implementation of MateStack class


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+
+    public boolean isEmpty() {
+        return size == 0;
+    }
+
+    private static class Node<T> {
+        private T element;
+        private Node<T> next;
+
+        public Node(T element, Node<T> next) {
```
slade13, line 46: Don’t overuse public access modifiers. Since Node is only used inside `MateStack`, make the constructor private (or omit public) to minimize exposure.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> head;
+    private int size;
 
+    public void push(T value) {
+        head = new Node<>(value,head);
```
slade13, line 10: You use value in `push` but `element` inside `Node`. Consider using the same term (value) everywhere for readability.
```suggestion
        head = new Node<>(value, head);
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,51 @@
...
+        size++;
     }
 
     public T peek() {
-        return null;
+        if (head == null) {
```
slade13, line 15: Extract a private helper to avoid repeating the empty check in `peek()` and `pop()`. This also demonstrates “keep utility methods as restricted as possible”.

src/main/java/core/stack/MateStack.java
slade13, line 10: Decide whether null values are allowed on the stack. Your current code supports pushing `null`. If that’s unwanted, add a check in push and throw `NullPointerException` (or document that `null` is allowed).

slade13 (approved):
Good job, please review my comments.

PR #181, opened 2025-09-26
My linked Stack

koder95:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,44 @@
...
+    }
+
+    private Node<T> top = null;
+    private int size = 0;
 
+    public void push(T value) {
+        Node<T> top = new Node<>(value);
+        top.prev = this.top;
+        this.top = top;
+        size++;
```
MateuszRostek, line 22: I would suggest creating and using constructor with all fields, something like this:

```suggestion
        Node<T> first = head;
        head = new Node<>(first, value);
        size++;
```

MateuszRostek (approved):
Good job! Check one small advice from my side + comments from ai-mentor

PR #183, opened 2025-10-01
Develop

Senjars:
Stack recreated

### DIFF
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -33,9 +33,9 @@ public class MateStack<T> {
             throw new EmptyStackException();
         }
         size--;
-        T value = array[size];
+        T result = array[size];
         array[size] = null;
-        return value;
+        return result;
     }
 
     public int size() {
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -1,19 +1,44 @@
 package core.stack;
 
-public class MateStack<T> {
-    public void push(T value) {
+import java.util.EmptyStackException;
 
+public class MateStack<T> {
+    private int size = 0;
+    private int arrayCapacity = 16;
+    private T[] array = (T[]) new Object[arrayCapacity];
+
+    public void push(T value) {
+        if (size >= arrayCapacity) {
+            arrayCapacity *= 2;
+            T[] newArray = (T[]) new Object[arrayCapacity];
+
+            for (int i = 0; i < array.length; i++) {
+                newArray[i] = array[i];
+            }
+            array = newArray;
+        }
+        size++;
+        array[size - 1] = value;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        return array[size - 1];
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+        size--;
+        T value = array[size];
+        array[size] = null;
+        return value;
     }
 
     public int size() {
-        return 0;
+        return size;
     }
 }
--- a/src/main/java/core/stack/MateStack.java
+++ b/src/main/java/core/stack/MateStack.java
@@ -12,7 +12,7 @@ public class MateStack<T> {
             arrayCapacity *= 2;
             T[] newArray = (T[]) new Object[arrayCapacity];
 
-            for (int i = 0; i < array.length; i++) {
+            for (int i = 0; i < size; i++) {
                 newArray[i] = array[i];
             }
             array = newArray;
@@ -40,5 +40,9 @@ public class MateStack<T> {
 
     public int size() {
         return size;
+    }
+
+    public boolean isEmpty() {
+        return size == 0;
     }
 }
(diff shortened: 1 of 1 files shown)


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
+    private int size = 0;
+    private int arrayCapacity = 16;
```
slade13: Extract this magic number to constant:
`private static final int DEFAULT_CAPACITY = 16;`

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+    private T[] array = (T[]) new Object[arrayCapacity];
+
     public void push(T value) {
+        if (size >= arrayCapacity) {
+            arrayCapacity *= 2;
+            T[] newArray = (T[]) new Object[arrayCapacity];
 
+            for (int i = 0; i < size; i++) {
```
slade13: To copy array use: `System.arraycopy(...)`

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
+    private int size = 0;
+    private int arrayCapacity = 16;
+    private T[] array = (T[]) new Object[arrayCapacity];
+
     public void push(T value) {
+        if (size >= arrayCapacity) {
```
slade13: Extract checking array size and increasing its size to separate method: `ensureCapacity(int minCapacity)`

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+            arrayCapacity *= 2;
+            T[] newArray = (T[]) new Object[arrayCapacity];
```
slade13: Use here `@SuppressWarnings("unchecked")`

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
+    private int size = 0;
+    private int arrayCapacity = 16;
+    private T[] array = (T[]) new Object[arrayCapacity];
```
slade13: Use here `@SuppressWarnings("unchecked")`

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+            }
+            array = newArray;
+        }
+        size++;
+        array[size - 1] = value;
     }
 
     public T peek() {
-        return null;
+        if (size == 0) {
```
slade13: Extract this to helper method.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
...
+            throw new EmptyStackException();
+        }
+        return array[size - 1];
     }
 
     public T pop() {
-        return null;
+        if (size == 0) {
```
slade13: Extract this to helper method.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,48 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
```
slade13, line 5: You can consider adding a constructor that accepts initial capacity and validates it.

src/main/java/core/stack/MateStack.java
slade13: Don’t track capacity in a separate arrayCapacity field — it can drift from array.length. Use array.length.

slade13 (changes requested):
Your solution is good in overall, but there are some nitty things that I would like you to change. Please change them and I'll review it again.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
+
+    @SuppressWarnings("unchecked")
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > array.length) {
+            int newCapacity = array.length * 2;
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+            }
+            T[] newArray = (T[]) new Object[newCapacity];
+            System.arraycopy(array,0,newArray,0,size);
```
slade13, line 55:
```suggestion
            System.arraycopy(array, 0, newArray, 0, size);
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,59 @@
...
+    private void ensureNotEmpty() {
+        if (size == 0) {
+            throw new EmptyStackException();
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > array.length) {
+            int newCapacity = array.length * 2;
```
slade13, line 50: 2 is magic number, in this small project it doesn't look like a big thing - but imagine working in large project where you spot some number and nobody know from where this value came from.

It's better to store it for example as  `GROWTH_FACTOR`

slade13 (approved):
Good job!

PR #184, opened 2025-10-10
Task Stack full impl with clean build

diego6703:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+        return removedNode.value;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private class Node<T> {
+        private T value;
```
mateuszwojtkowiak, line 43: 'value' variable should also be final.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+        size--;
+        return removedNode.value;
     }
 
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private class Node<T> {
```
mateuszwojtkowiak, line 42: Class Node should also be static.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+
 public class MateStack<T> {
-    public void push(T value) {
 
+    private Node<T> head;
+    private int size;
+
+    public void push(T value) {
+        Node<T> newNode = new Node<>(value);
+        if (head == null) {
```
mateuszwojtkowiak, line 12: We can make this method much shorter if we don't use conditional statements: 
public void push(T value) {
        Node<T> newNode = new Node<>(value);
        newNode.prev = head; 
        head = newNode;
        size++;
    }

mateuszwojtkowiak (approved):
Great job! Your code is pretty good and there is not much to fix. You can read my comments and see what can be improved.

PR #197, opened 2026-01-02
Solution

Rabico:


src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail = new Node<>(null, null);
```
slade13: `tail `is acceptable; many stacks call it `top`

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
 
     public int size() {
-        return 0;
+
+        return arrSize;
+    }
+
+    class Node<T> {
+        private Node<T> prev;
+        private T item;
```
slade13: `item `is okay, but `value `or `data `is also fine.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
 package core.stack;
 
+import java.util.EmptyStackException;
+
 public class MateStack<T> {
-    public void push(T value) {
+    private Node<T> tail = new Node<>(null, null);
+    private int arrSize = 0;
```
slade13: `arrSize` is misleading (this is not an array). Prefer `size`.
```suggestion
    private int size = 0;
```

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+        }
+        arrSize--;
+        T value = tail.item;
+        tail = tail.prev;
+        return value;
     }
 
     public int size() {
-        return 0;
+
```
slade13: Remove this empty line.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+    private int arrSize = 0;
 
+    public void push(T value) {
+        if (arrSize == 0) {
+            tail.item = value;
+        } else {
+            tail = new Node<>(tail, value);
+        }
+        arrSize++;
     }
```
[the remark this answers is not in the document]
slade13: Just like AI mentor wrote, there is an bug, you have to fix it. Initialize `tail `with `null `and change your code accordingly.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
-        return 0;
+
+        return arrSize;
+    }
+
+    class Node<T> {
+        private Node<T> prev;
+        private T item;
+
+        public Node(Node<T> prev, T item) {
```
slade13: Your `Node `constructor is `public`. It’s only used inside the outer class → make it `private`.

src/main/java/core/stack/MateStack.java
```diff
@@ -1,19 +1,50 @@
...
+        return value;
     }
 
     public int size() {
-        return 0;
+
+        return arrSize;
+    }
+
+    class Node<T> {
```
slade13: Your inner class `Node<T>` has **default access**. Make it:

```suggestion
    private static class Node<T> {
```

slade13 (changes requested):
Good job, but your code still needs some changes, please them and send your code again for a review. Besides that your commit messages should be descriptive and should be in line with the changes which you provide in such commit. Messages like you have now:

> solution
> solution
> solution 2
> solution 3
> solution 4

are unacceptable. Please change these messages using git interactive rebase: https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase

MateuszRostek (approved):

